PR #8, opened 2025-02-01
fix(openssl fips): correct openssl fips package names in checks and comments

stevebeattie:
- Fix the fips openssl package install checks to be more tightly bound, both to the record type in the apk install db in `/lib/apk/db/installed` as well as to the package names themselves, specifically look for openssl-config-fipshardened and openssl-provider-fips or the version-stream variant thereof.
- Fix comment and description references to refer to the (non-version-stream) correct package names
- Fix several other textual tyops found by codespell as well


xnox (commented):
Looks good! And it is future proof for upcoming certifications, which will help in the future.

I will run these against a few containers to validate, and then merge.

We really need sample unit-test here (skeleton FIPS image possibly) to check these.

xnox (approved):
Passes against fips containers; and correctly failed against regular containers.

PR #13, opened 2025-05-27
Hash script to oval check change

jamie-albert:
This needs to be an oval check likely, see here for more details: https://github.com/chainguard-dev/prodsec/issues/189


stevebeattie (approved):
Thanks, I've confirmed that the oval check works correctly, that the test for file existence is also detecting correctly, and also that the updated hashsum is correct.

PR #20, opened 2025-07-16
Update ID mappings to fix STIGViewer imports

egibs:
This PR updates our GPOS STIG and Benchmark IDs to allow for the following STIGViewer workflow:
- `File` -> `Import STIG`
- Select `ssg-chainguard-gpos-ds.xml` (or import `U_GPOS_V3R2_SRG.zip` directly)
- `Checklist` -> `Create Checklist - Check Marked STIG(s)`
- `Import XCCDF Results File` -> `results.xml`

For the most part, this follows existing STIG patterns and seems to work without issue (plus the original `results.html` file still displays as it always did).

image


gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -6838,7 +7412,7 @@
             <states>
             <ind:filehash58_state id="oval:org.CABundleHash:ste:1" version="1">
                 <ind:hash_type>SHA-256</ind:hash_type>
-                <ind:hash>71ad3cc35a5f1b5ba09256a997afc827885925a701c91432ea3fb20201664fb0</ind:hash>
+                <ind:hash>8cf3f2d5222508ca469c78a347764ee1b1fab10a903172612c489ebcca05c91c</ind:hash>
```
egibs, line 7415: This is the only non-ID change.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -188,238 +189,264 @@
...
+                            changes, passphrase-length rules, or password-generation aids.
+
+                            For more information on inherited controls for containers see the DISA
+                            Container Hardening Process Guide.</ns0:rationale>
+                        <ns0:platform idref="cpe:/o:notapplicable" />
+                    </ns0:Rule>
+                </ns0:Group>
+                                <ns0:Group id="xccdf_mil.disa.stig_group_V-263654">
+    <ns0:Rule id="xccdf_mil.disa.stig_rule_SV-263654r1_rule" selected="true" severity="medium">
+                    <ns0:version>SRG-OS-263654-GPOS-263654</ns0:version>
```
stevebeattie, line 229: Unfortunately, I don't this is quite right. The vulnerability id is not directly mapped into the SRG version id; e.g. this is for `V-263654` which maps to this: https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000720-GPOS-00170/ so I *think* the `ns0:version` ought to be `SRG-OS-000720-GPOS-00170`

egibs: Nice, good catch. I'll get this updated.

egibs: Updated in [`e78d74e` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/e78d74e0a2ab49c4884ab06fcc50cb44cb8f00fe).

egibs:
Fixed the validation for the files in https://github.com/chainguard-dev/stigs/pull/20/commits/53b25596b7d34a263dda2d97f43447b91d683fde and https://github.com/chainguard-dev/stigs/pull/20/commits/1a6dbad66cef763435cf52a6a296850f087f6e16:
```
$ docker run -i --rm -u 0:0 --pid=host -v /var/run/docker.sock:/var/run/docker.sock   -v $(pwd)/out:/out -v $(pwd):/workspace cgr.dev/chainguard/openscap:latest-dev ds sds-validate /workspace/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
$
$ $ docker run -i --rm -u 0:0 --pid=host -v /var/run/docker.sock:/var/run/docker.sock -v $(pwd)/out:/out -v $(pwd):/workspace cgr.dev/chainguard/openscap:latest-dev xccdf validate /workspace/gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
<?xml version="1.0"?>
Warning: The 'cpe:/' prefix (CPE URI binding) is allowed within an @idref attribute, but the CPE Formatted String binding is preferred. See the XCCDF 1.2.1 specification, Section 6.2.5.

File: gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml [hunk 3]
```diff
@@ -167,7 +167,7 @@
             </ns0:Profile>
             <ns0:Group id="xccdf_mil.disa.stig_group_V-263651">
                 <ns0:Rule id="xccdf_mil.disa.stig_rule_SV-263651r1_rule" selected="true" severity="medium">
-                    <ns0:version>SRG-OS-263651-GPOS-263651</ns0:version>
+                    <ns0:version>SRG-OS-000690-GPOS-00140</ns0:version>
```
stevebeattie: I haven't validated the vast majority of these, but spot checking a few, they look correct.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -166,7 +167,10 @@
 
             </ns0:Profile>
             <ns0:Group id="xccdf_mil.disa.stig_group_V-263651">
-                <ns0:Rule id="xccdf_mil.disa.stig_rule_SV-263651r1_rule" selected="true" severity="medium">
+        <ns0:title>Security Group - xccdf_mil.disa.stig_group_V-263651</ns0:title>
+                <ns0:title xml:lang="en">SRG-OS Group V-263651</ns0:title>
+                <ns0:description xml:lang="en">&lt;GroupDescription&gt;&lt;/GroupDescription&gt;</ns0:description>
+                <ns0:Rule id="xccdf_mil.disa.stig_rule_SV-263651r1_rule" selected="true" severity="medium" weight="10.0">
```
stevebeattie: The rules (encoded as SV-NNNNNrREVISION) have a more complex revision number than "r1" in the DISA reference xccdf, e.g. for `V-263651`, the related SV rule id is `SV-263651r982555_rule`.

The added group description here doesn't seem useful, since it's not got any content.

Adding in the weight field I guess is fine, in the DISA STIG they all seem to be 10.0.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -188,9 +192,20 @@
                         For more information on inherited controls for containers see the DISA
                         Container Hardening Process Guide.</ns0:rationale>
                     <ns0:platform idref="cpe:/o:notapplicable" />
+                    <ns0:ident system="http://cyber.mil/legacy">V-263651</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/legacy">SV-263651</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/cci">CCI-000366</ns0:ident>
```
stevebeattie: The same CCI (`CCI-000366`) is used throughout. They are different per Vuln/Rule; for example, the one for `V-263651` here is `CCI-003959` (https://stigaview.com/products/srg-gpos/v3r2/SRG-OS-000690-GPOS-00140/)

There's a whole lookup table on what CCI mapping are: https://public.cyber.mil/stigs/cci/

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -222,11 +237,22 @@
                             For more information on inherited controls for containers see the DISA
                             Container Hardening Process Guide.</ns0:rationale>
                         <ns0:platform idref="cpe:/o:notapplicable" />
+        <ns0:ident system="http://cyber.mil/legacy">V-263653</ns0:ident>
+        <ns0:ident system="http://cyber.mil/legacy">SV-263653</ns0:ident>
+        <ns0:ident system="http://cyber.mil/cci">CCI-000366</ns0:ident>
+        <ns0:fixtext fixref="F-263653r1_fix">Configure the operating system to meet the requirement.</ns0:fixtext>
+        <ns0:fix id="F-263653r1_fix" />
```
stevebeattie: I'm not sure about the generic fix/remediation text, especially for the rules we've marked as not applicable, and thus have no check.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -188,9 +192,20 @@
...
                         Container Hardening Process Guide.</ns0:rationale>
                     <ns0:platform idref="cpe:/o:notapplicable" />
+                    <ns0:ident system="http://cyber.mil/legacy">V-263651</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/legacy">SV-263651</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/cci">CCI-000366</ns0:ident>
+                    <ns0:fixtext fixref="F-263651r1_fix">Configure the operating system to meet the requirement.</ns0:fixtext>
+                    <ns0:fix id="F-263651r1_fix" />
+                    <ns0:check system="http://oval.mitre.org/XMLSchema/oval-definitions-5">
+                        <ns0:check-content-ref href="stub-oval.xml" />
+                    </ns0:check>
```
stevebeattie: I'm not sure why these stub-oval checks got added; looking quickly, they seem to be added for the rules that we've marked ass not-applicable, and so shouldn't have a check, I think. (oscap warns about it when evaluating the xccdf against an image.)

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -6662,13 +8507,20 @@
...
-                    <ns0:check system="http://open-scap.org/page/SCE">
-                        <ns0:check-import import-name="stdout" />
-                        <ns0:check-content-ref href="./AslrCheck.sh" />
+                    <ns0:ident system="http://cyber.mil/legacy">V-203754</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/legacy">SV-203754</ns0:ident>
+                    <ns0:ident system="http://cyber.mil/cci">CCI-000366</ns0:ident>
+                    <ns0:fixtext fixref="F-203754r1_fix">Configure the operating system to meet the requirement.</ns0:fixtext>
+                    <ns0:fix id="F-203754r1_fix" />
+                    <ns0:check system="http://oval.mitre.org/XMLSchema/oval-definitions-5">
+                        <ns0:check-content-ref href="stub-oval.xml" />
```
stevebeattie: This drops the check that address space layout randomization is enabled entirely, and converts it to the stub-oval check, and causes `notchecked` results in oscap, and Not Reviewed results in STIG Viewer when imported there.

(One wrinkle is that the script that does the check needs to happen on a running system, because it's checking a sysctl value exported from the running kernel's `/proc/sys` interface. So we should encourage checking against a running container over a given image, I suppose.)

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" ?>
-<Benchmark xmlns="http://checklists.nist.gov/xccdf/1.2" xmlns:html="http://www.w3.org/1999/xhtml" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="xccdf_oval_benchmark_.beta" style="SCAP_1.2" resolved="false" xml:lang="en-US" xsi:schemaLocation="http://checklists.nist.gov/xccdf/1.2 xccdf-1.2.xsd">
+<Benchmark xmlns="http://checklists.nist.gov/xccdf/1.2" xmlns:html="http://www.w3.org/1999/xhtml" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:cpe2="http://cpe.mitre.org/language/2.0" id="xccdf_oval_benchmark_.beta" style="SCAP_1.2" resolved="false" xml:lang="en-US" xsi:schemaLocation="http://checklists.nist.gov/xccdf/1.2 xccdf-1.2.xsd http://cpe.mitre.org/language/2.0 cpe-language_2.0.xsd">
   <status date="2023-04-01">draft</status>
   <title>XCCDF Checklist Beta</title>
   <description>This XCCDF document provides a basic security check for a Linux system.</description>
   <notice id="legal">Copyright (c) 2023, Your Organization. All rights reserved.</notice>
   <reference href="http://yourorganization.com/security-policies">Security Policy Reference</reference>
+  <platform idref="cpe:/o:linux:linux_kernel"/>
```
stevebeattie, line 8: not sure this is correct, especially since our containers are not shipping a kernel.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
```diff
@@ -18,7 +19,7 @@
     <dc:identifier>Chainguard-GPOS-OVAL</dc:identifier>
     <dc:language>en-US</dc:language>
   </metadata>
-  <platform idref="cpe:/o:chainguard:chainguard_linux"/>
+  <model system="urn:xccdf:scoring:default"/>
```
stevebeattie, line 22: ... and also, I don't think we should drop the chainguard specific cpe... though searching NIST's CPE database, I don't see a specific chainguard or wolfi cpe filed, so I guess we should ask for one.

stevebeattie (changes requested):
Ugh, github won't let me comment inline on the unified diff of all commits, so I'm having to add comments on the individual commits, sorry about that. We're getting better, but there are still some issues.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
egibs: Fixed in: [`35c8b42` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/35c8b42d08dabf4f04c912a8a698ee9fd7a423d6)

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
egibs: Fixed in: [`35c8b42` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/35c8b42d08dabf4f04c912a8a698ee9fd7a423d6)

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
egibs: Fixed in: [`35c8b42` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/35c8b42d08dabf4f04c912a8a698ee9fd7a423d6)

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -71,102 +71,103 @@
...
+                <ns0:select idref="xccdf_mil.disa.stig_rule_SV-203716r1_rule" selected="true" />
+                <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263650r1_rule" selected="true" />
+                <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263652r1_rule" selected="true" />
+                <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263659r1_rule" selected="true" />
 
 
             </ns0:Profile>
-            <ns0:Group id="xccdf_._group_not_applicable_tests">
-                <ns0:Rule id="xccdf_._rule_V_263651" selected="true" severity="medium">
+            <ns0:Group id="xccdf_mil.disa.stig_group_V-263651">
```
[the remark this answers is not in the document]
egibs: Updated in: [`35c8b42` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/35c8b42d08dabf4f04c912a8a698ee9fd7a423d6)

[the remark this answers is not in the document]
egibs: e.g.,
image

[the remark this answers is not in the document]
egibs: This evaluates correctly against the upstream STIG as well:
image

egibs:
I can import the results cleanly against both STIGs with the latest changes. The upstream STIG contains more detail, but since the results work with either it's trivial to swap between them.

Using CATI as an example --
Our STIG:
CleanShot 2025-07-18 at 10 15 52@2x

Upstream (straing from the `.zip` archive):
CleanShot 2025-07-18 at 10 16 39@2x

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -6878,7 +6997,8 @@
                     id="oval:org.RemoteAccessServices:obj:6" version="1">
                     <path>/lib/apk/db</path>
                     <filename>installed</filename>
-                    <pattern operation="pattern match">^P:(openssh|openssh-server|openssh-client|openssh-sftp-server|dropbear|tigervnc|tigervnc-server|tigervnc-viewer|xrdp|xorgxrdp|vsftpd|proftpd|webmin|cockpit|cockpit-ws|cockpit-bridge|nfs-utils|samba|samba-server|samba-client|samba-common|rsh|telnet)-*</pattern>
+                    <pattern operation="pattern match">
+                        ^P:(openssh|openssh-server|openssh-client|openssh-sftp-server|dropbear|tigervnc|tigervnc-server|tigervnc-viewer|xrdp|xorgxrdp|vsftpd|proftpd|webmin|cockpit|cockpit-ws|cockpit-bridge|nfs-utils|samba|samba-server|samba-client|samba-common|rsh|telnet)-*</pattern>
```
stevebeattie, line 7001: The pattern match cannot be moved to a new line, as the addition of the newline will become part of the patte to match, and thus won'tever match, rendering the test always true.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
```diff
@@ -6021,22 +6124,23 @@
...
-
-<html:pre>Script Verification:
-                            To manually verify, run command 'cat /etc/shadow and verify that no
-                            additional users exist</html:pre></ns0:description>
+                        logging that can be used for forensic purposes. To address access
+                        requirements, many operating systems can be integrated with enterprise-level
+                        authentication/access/auditing mechanisms that meet or exceed access control
+                        policy requirements.
+
+                        Script Verification:
```
stevebeattie, line 6132: I think the reason our stig had the pre-formatted text was to make it stand out in an oscap html report, like so: 
Screenshot from 2025-07-20 21-56-14

But I'm not sure this is the proper way to express manual steps like this, or if there's a separate tag instead of `description` that should be used.

stevebeattie (changes requested):
One more thing to note is that, with the rule id changes in place, we'll have some tests in a couple of packages to update; https://github.com/wolfi-dev/os/blob/main/chainguard-security-guide.yaml being one of them.

gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
egibs: Yea, I was having all sorts of issues with that. I'll see about fixing that.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalDefinitions/DetectOpenSslTest.xml
```diff
@@ -11,62 +14,61 @@
...
-    <textfilecontent54_object xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" id="oval:org.OpenSsl:obj:4" version="1">
-      <path>/lib/apk/db</path>
-      <filename>installed</filename>
-      <pattern operation="pattern match">^P:openssl-provider-fips(-\d+\.\d+\.\d+)?$</pattern>
-      <instance datatype="int">1</instance>
-    </textfilecontent54_object>
+    <independent:textfilecontent54_object id="oval:org.OpenSsl:obj:4" xmlns:independent="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" version="1">
+      <independent:path>/lib/apk/db</independent:path>
+      <independent:filename>installed</independent:filename>
+      <independent:pattern operation="pattern match">^P:openssl-provider-fips(-[0-9]+\.[0-9]+\.[0-9]+)?$</independent:pattern>
```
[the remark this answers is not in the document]
egibs: Reverted in [`2ae1599` (#20)](https://github.com/chainguard-dev/stigs/pull/20/commits/2ae15996b940b888e068bb210bca47ab84c95759).

stevebeattie (approved):
Verified that I can use oscap/oscap-docker to generate results, that the html report looks sensible in both success and failure conditions, that the results files can be imported into a stig checklist with both stig viewer v2 (when applied to both the disa xccdf and our xccdf) and v3.

Thanks for all your efforts!

brooksphilip:
Can we add one more change to output the test run, the result, and the pass/fail to the finding details box?

PR #23, opened 2025-10-14
Update CABundleHash

egibs:
I need to look into why the automation hasn't picked this up yet, but this will address the drift in the meantime.


some-natalie (approved):
Thank you for bumping this CA bundle SHA!

I'm curious if there's an easy way to bump this XML element in this file using octo-sts to remove the human from keeping it in sync ... 🤔

PR #70, opened 2026-03-31
fix RemoteAccessServicesTest: version-stream-aware pattern

jakeva:
The `-*` suffix in the package pattern was doing nothing useful (zero or more literal dashes), and without a `$` anchor, packages like `samba-libs` and `samba-util-libs` matched `samba` as a prefix, causing false-positive No_Remote_Access failures.

Replaces the suffix with `(-\d[\d.]*(-[a-z][a-z0-9-]*)?)?$` to correctly handle version-streamed variants (`openssh-9.4`, `samba-4.19-client`) while rejecting alphabetic-only suffixes like `-libs` or `-keysign`.


stevebeattie (dismissed):
It's correct to note that with the lack of the trailing anchor, the pattern will match more things than just the packages listed; however the proposed fix will match those packages and superfluously  trailing `-`s.

Because we do have version-streamed packages, we should probably adjust the regex to look for version-streamed variants. That said, the existing regex did not do this correctly either (e.g. wouldn't match `mysql-9.4-client`).

stevebeattie (commented):
Thanks, nice fix.

stevebeattie (approved):

PR #116, opened 2026-06-26
feat: add offline test harness

egibs:
## What

A Go test harness (`tests/oscap-offline/`) that scans a **fixture filesystem tree**
with OpenSCAP by pointing the scanner at an **alternate root** via `OSCAP_PROBE_ROOT`,
then asserts **per-rule** pass/fail. It mirrors the chainguard-inspec pattern (lay out a
tree → scan it as if it were `/` → check one rule), applied to this repo's XCCDF/OVAL
datastream.

- Fixtures are `base wolfi-base + overlay mutations`.
- Coverage is a table-driven **per-OVAL-definition pass+fail matrix** — adding a new
  fixture is a one-row change.
- It runs **alongside** (not replacing) the existing `oscap-docker` e2e tier.

## Why Go

The harness is built in Go to take advantage of capabilities that map directly onto what
this kind of testing needs:

- **Structured result handling.** XCCDF results are parsed with `encoding/xml` into typed
  structs, and assertions are made per rule from the parsed report. This keeps the
  pass/fail logic explicit and independent of the scanner's process exit code (where
  `0` and `2` are both valid).
- **First-class parallelism and fixture discovery.** Fixtures are table-driven test cases
  run with `t.Parallel`, so the suite scales to many fixtures with built-in concurrency
  and a single place to add a case.
- **In-process image export.** The base rootfs is exported with
  `go-containerregistry` (`crane.Export`) rather than a container build, which keeps the
  export hermetic and aligns with the org's go-standards guidance to use library calls for
  registry operations.
- **Syscall-level extraction confinement.** Extraction uses Go's `os.Root`
  (openat2/`RESOLVE_BENEATH` on Linux), which confines all filesystem operations beneath
  the destination directory at the OS level. The extractor is a single, audited primitive
  (`safetar`) with unit and fuzz coverage, and the tar-producing side is validated with
  `filepath.IsLocal`.
- **Standard tooling and review surface.** The module follows the org go-standards (layout,
  `doc.go`/`example_test.go`, `gofmt`, `golangci-lint` with `gosec`), giving consistent
  linting, formatting, and static analysis.

## Why `OSCAP_PROBE_ROOT` (vs. `oscap-docker` per fixture)

`OSCAP_PROBE_ROOT` is the direct analog of inspec's `rootfs` input: it scans an arbitrary
directory tree as the filesystem root, with no image build, no docker socket, and no
privileged scan per fixture. A feasibility spike confirmed the mechanism — mutating the
tree changes what `oscap` reads and flips a rule from pass to fail — so the design rests on
a verified behavior.

## Notes / caveats

- **Ownership model:** fixture file uid/gid/mode are set in **tar headers** (so no host
  root is required to build a fixture), and extraction runs as **root inside the scan
  container** so the OVAL `file` probe observes real `root:root`. This matters because
  `VarLogPermissions`/`LibraryPermissions` assert `uid==0`, not mode bits.
- **Container surface:** no docker socket; `/src` and fixtures mounted `:ro`; only `/out`
  writable; no `--privileged`.
- **SCE rules excluded:** the 2 `AslrCheck.sh`/`sysctl` rules read the live host kernel,
  not the offline root, so they are intentionally outside the matrix — enforced by a test.
- **`DetectOpenSsl` is inverted** (the OVAL *requires* the FIPS module present), confirmed
  against the OVAL definition.
- **Determinism:** the harness reads the *same* pinned wolfi-base digest the fixtures use,
  so `CertificateAudit`'s hash check stays green; `update-ca-cert.yaml` keeps that pin in
  lockstep.
- **Two tiers retained:** the fast offline harness complements the existing heavier
  `oscap-docker` integration tier rather than replacing it.


tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+}
+
+// Synthetic mutation content. Each blob is the minimal change that flips one
+// OVAL definition's verdict; the comment cites the pattern/state it satisfies.
+
+// apkOpenSSHStanza is an apk installed-db record whose P: line matches the
+// RemoteAccessServices pattern (openssh-server is in the banned package list),
+// so the none_exist test fails.
+var apkOpenSSHStanza = []byte("\nP:openssh-server\nV:9.9_p2-r0\nA:x86_64\n")
+
```
stevebeattie, line 284:
```suggestion

// apkOpenSSHKeygenStanza records openssh-keygen, whose "-keygen" word suffix the
// RemoteAccessServices pattern `^P:(...openssh...)(-\d[\d.]*(-[a-z][a-z0-9-]*)?)?$`
// must NOT match: the optional version suffix requires a leading digit, so a
// word-suffixed sibling of a banned package falls through and the rule PASSES.
// Guards the boundary against a regex regression that drops the `$` anchor or the
// digit class and starts flagging legitimate packages.
var apkOpenSSHKeygenStanza = []byte("\nP:openssh-keygen\nV:9.9_p2-r0\nA:x86_64\n")

```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+// plainHTTPRepo is a non-https, non-comment repository line, which the
+// PackageSignature none_exist pattern `^(?!\s*#)(?!.*https://).+$` matches.
+var plainHTTPRepo = []byte("http://insecure.example.com/alpine\n")
+
```
stevebeattie, line 296:
```suggestion

// commentedAndHTTPSRepos are two repository lines the PackageSignature pattern
// `^(?!\s*#)(?!.*https://).+$` must NOT match: a comment line is exempt via the
// `(?!\s*#)` branch and an https line via the `(?!.*https://)` branch. Appending
// both to a clean base must keep the rule PASSING — guarding the two exemption
// branches that the single plain-http fail fixture never exercises.
var commentedAndHTTPSRepos = []byte("# http://insecure.example.com/alpine\nhttps://packages.example.com/alpine\n")

```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+// caTamper appends content to the CA bundle so its SHA-256 diverges from the
+// datastream's pinned hash, failing the filehash58 state.
+var caTamper = []byte("\n# tamper\n-----BEGIN CERTIFICATE-----\nTAMPERED\n-----END CERTIFICATE-----\n")
+
+// activeShadowEntry is an /etc/shadow line whose password field is a
+// traditional DES crypt hash (not "!" or "*"), which the UserPasswordConfigured
+// pattern `^[^:]+:(?![!*])[^:\n]*:` matches, failing the none_exist test.
+var activeShadowEntry = []byte("compliance-test:ZZx/p0vU8jVbA:19000:0:99999:7:::\n")
+
```
stevebeattie, line 312:
```suggestion

// emptyShadowPassword is an /etc/shadow line whose password field is EMPTY
// (passwordless login), distinct from a hashed or a locked ("!"/"*") field. The
// UserPasswordConfigured pattern `^[^:]+:(?![!*])[^:\n]*:` still matches it — the
// field neither starts with "!"/"*" nor is otherwise exempt — so the none_exist
// test must FAIL. This guards the negative-lookahead's handling of the empty
// field, the case a hashed-password fixture never exercises.
var emptyShadowPassword = []byte("emptyacct::19000:0:99999:7:::\n")

```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+// fipsModuleCnf, opensslCnf, and apkFIPSPackages together satisfy all seven
+// DetectOpenSsl criteria: the two config files must exist, both FIPS packages
+// must be recorded in the apk db, and openssl.cnf must contain the include line,
+// the provider_sect fips line, and the default_properties fips=yes line.
+var (
+	fipsModuleCnf   = []byte("[fips_sect]\nactivate = 1\n")
+	opensslCnf      = []byte(".include fipsmodule.cnf\n[provider_sect]\nfips = fips_sect\ndefault = default_sect\n[algorithm_sect]\ndefault_properties = fips=yes\n")
+	apkFIPSPackages = []byte("\nP:openssl-config-fipshardened\nV:3.5.1-r0\nA:x86_64\n\nP:openssl-provider-fips\nV:3.5.1-r0\nA:x86_64\n")
+)
+
```
stevebeattie, line 335:
```suggestion

// apkFIPSPackagesDocSubpkg records only the doc subpackages
// (openssl-config-fipshardened-doc / openssl-provider-fips-doc). The DetectOpenSsl
// package patterns `^P:openssl-...(-\d+\.\d+\.\d+)?$` exclude word suffixes, so
// these must NOT satisfy the package criterion. Paired with the config files
// present, the rule must stay FAIL — isolating the package-pattern branch so a
// regex that accepts word-suffixed subpackages would be caught.
var apkFIPSPackagesDocSubpkg = []byte("\nP:openssl-config-fipshardened-doc\nV:3.5.1-r0\nA:x86_64\n\nP:openssl-provider-fips-doc\nV:3.5.1-r0\nA:x86_64\n")

```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+		// RemoteAccessServices: clean base has no banned remote-access package.
+		{
+			name: "remote_access/pass_clean",
+			want: map[string]results.Result{ruleRemoteAccessServices: results.Pass},
+		},
+		{
+			name: "remote_access/fail_openssh_installed",
+			ops:  []overlay.Op{overlay.AppendFile("usr/lib/apk/db/installed", apkOpenSSHStanza)},
+			want: map[string]results.Result{ruleRemoteAccessServices: results.Fail},
+		},
```
stevebeattie, line 378:
```suggestion
		},
		{
			// openssh-keygen shares a banned prefix but its "-keygen" word suffix
			// must not match the anchored version-suffix pattern, so the rule passes.
			name: "remote_access/pass_word_suffix_not_banned",
			ops:  []overlay.Op{overlay.AppendFile("usr/lib/apk/db/installed", apkOpenSSHKeygenStanza)},
			want: map[string]results.Result{ruleRemoteAccessServices: results.Pass},
		},
```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+		// PackageSignature: clean base ships only https apk repositories.
+		{
+			name: "package_signature/pass_clean",
+			want: map[string]results.Result{rulePackageSignature: results.Pass},
+		},
+		{
+			name: "package_signature/fail_plain_http_repo",
+			ops:  []overlay.Op{overlay.AppendFile("etc/apk/repositories", plainHTTPRepo)},
+			want: map[string]results.Result{rulePackageSignature: results.Fail},
+		},
```
stevebeattie, line 396:
```suggestion
		},
		{
			// A comment line and an https line are both exempt from the pattern, so
			// appending them to the clean https-only base keeps the rule passing.
			name: "package_signature/pass_comment_and_https",
			ops:  []overlay.Op{overlay.AppendFile("etc/apk/repositories", commentedAndHTTPSRepos)},
			want: map[string]results.Result{rulePackageSignature: results.Pass},
		},
```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+		// UserPasswordConfigured: clean base has only locked ("!"/"*") accounts.
+		{
+			name: "user_password/pass_clean",
+			want: map[string]results.Result{ruleUserPasswordConfigured: results.Pass},
+		},
+		{
+			name: "user_password/fail_active_password",
+			ops:  []overlay.Op{overlay.AppendFile("etc/shadow", activeShadowEntry)},
+			want: map[string]results.Result{ruleUserPasswordConfigured: results.Fail},
+		},
```
stevebeattie, line 427:
```suggestion
		},
		{
			// An empty password field is passwordless login; the negative-lookahead
			// still matches it, so the rule must fail.
			name: "user_password/fail_empty_password",
			ops:  []overlay.Op{overlay.AppendFile("etc/shadow", emptyShadowPassword)},
			want: map[string]results.Result{ruleUserPasswordConfigured: results.Fail},
		},
```

tests/oscap-offline/internal/scan/fixtures_test.go
```diff
@@ -0,0 +1,622 @@
...
+		{
+			name: "detect_openssl/pass_fips_present",
+			ops: []overlay.Op{
+				overlay.AddFile("etc/ssl/fipsmodule.cnf", fipsModuleCnf, 0o644, 0, 0),
+				overlay.AddFile("etc/ssl/openssl.cnf", opensslCnf, 0o644, 0, 0),
+				overlay.AppendFile("usr/lib/apk/db/installed", apkFIPSPackages),
+			},
+			want: map[string]results.Result{ruleDetectOpenSsl: results.Pass},
+		},
```
stevebeattie, line 485:
```suggestion
		},
		{
			// Config files present but only the *doc* subpackages installed: the
			// package patterns reject word suffixes, so the rule stays FAIL.
			name: "detect_openssl/fail_doc_subpackage_only",
			ops: []overlay.Op{
				overlay.AddFile("etc/ssl/fipsmodule.cnf", fipsModuleCnf, 0o644, 0, 0),
				overlay.AddFile("etc/ssl/openssl.cnf", opensslCnf, 0o644, 0, 0),
				overlay.AppendFile("usr/lib/apk/db/installed", apkFIPSPackagesDocSubpkg),
			},
			want: map[string]results.Result{ruleDetectOpenSsl: results.Fail},
		},
```

stevebeattie (dismissed):
Reviewed for correctness, security posture (these run on PRs), test design, and coverage against the chainguard-inspec suite. This is excellent work — `os.Root`-confined extraction with unit + fuzz coverage, hermetic in-process export, disciplined `errors.Is` usage, descriptive table-driven subtests, and a clean one-row-per-definition matrix. The SCE/ASLR exclusion and the uid-only (not mode) permission checks are correctly scoped and enforced. Approving.

The inline suggestions below are **optional and non-blocking** — four boundary fixtures that exercise OVAL-regex edges the existing happy-path pass/fail pairs don't: word-suffixed package names (`openssh-keygen`), comment/`https` repo exemptions, an empty-password shadow field, and `-doc` subpackage rejection. I verified all four locally against the pinned wolfi-base + `openscap:latest-dev` (`OSCAP_OFFLINE_REQUIRE=1`, race on) — each produces the asserted verdict.

One caveat worth a look: a dropped matrix row won't fail compilation (unused package-level vars are legal in Go), so the row count is the only guard against a fixture silently going missing.

stevebeattie (approved):

PR #125, opened 2026-07-17
refactor(oscap): fold SSL_CERT_FILE check into the CA-bundle rule (SV-263659)

xnox:
## What

Folds the `SSL_CERT_FILE` environment-variable assertion (added in #122)
into the existing CertificateAudit rule (`SV-263659`) instead of shipping
it as a separate rule.

`SV-263659` now passes only when **all three** hold:
1. the CA bundle exists,
2. its SHA-256 matches the datastream's pinned hash, and
3. `SSL_CERT_FILE` (from the image/container `Config.Env`) is set to
   `/etc/ssl/certs/ca-certificates.crt`.

## Datastream changes

- Add the `environmentvariable58` test as a third `AND` criterion of the
  `CABundleHash` OVAL definition; the test/object/state move into the
  `CABundleHash` namespace (`tst:3` / `obj:3` / `ste:2`).
- Remove the separate `SslCertFileEnv` definition, the
  `xccdf_com.chainguard_rule_ssl_cert_file_env` rule, and its profile
  `<select>`; revert the CA rule's `check-content-ref` to no `name=`
  (single definition again).
- Update the rule and OVAL descriptions to mention the SSL_CERT_FILE
  assertion.

## Offline matrix changes

- Drop the standalone `ssl_cert_file/*` cases.
- `certificate_audit/pass_clean` now also sets `SSL_CERT_FILE`.
- Add `certificate_audit/fail_wrong_ssl_cert_file` and
  `certificate_audit/fail_unset_ssl_cert_file` (valid bundle, bad env) to
  isolate the new dimension.

## Testing

- `oscap ds sds-validate`: PASS.
- Folded rule: clean+correct env -> pass; clean+wrong env -> fail;
  clean+unset env -> fail.
- Full offline module green, including all four `certificate_audit` cases.

## Note

Because the assertion is now folded into a DISA rule, `SV-263659` reports
a single verdict — a CA-bundle problem and an `SSL_CERT_FILE` problem are
no longer distinguishable in the report.

### DIFF
--- a/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
+++ b/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
@@ -133,7 +133,6 @@
         <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263650r982553_rule" selected="true"/>
         <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263652r982557_rule" selected="true"/>
         <ns0:select idref="xccdf_mil.disa.stig_rule_SV-263659r982563_rule" selected="true"/>
-        <ns0:select idref="xccdf_com.chainguard_rule_ssl_cert_file_env" selected="true"/>
       </ns0:Profile>
       <ns0:Group id="xccdf_._group_not_applicable_tests">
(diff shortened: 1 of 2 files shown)


xnox:
basically I don't like how the stig report looks with an extra check in the group; with a chainguard id; without a mapping to the GPOS id. Thus folding the check into an existing GPOS id for the rule.

egibs (approved):

PR #127, opened 2026-07-17
fix(oscap): anchor every rule to its OVAL definition id so reports show test details

xnox:
## What

Makes the XCCDF HTML report render the "OVAL test results details"
section for **every** OVAL-backed rule, not just the CA-bundle rule.

## Why

XCCDF rules backed by OVAL referenced their check document with a bare
`href` and no `name`. The report anchors each rule's OVAL detail section
off that `name` (the OVAL definition id), so with it absent the report
showed only a pass/fail verdict for those checks — never *what* was
tested. Only the CA-bundle rule rendered details, because it already
carried `name="oval:org.CABundleHash:def:1"`.

## Change

Add `name="<definition id>"` to all 88 OVAL `check-content-ref`s. Each id
is derived from the datastream's own catalog -> component -> definition
mapping (not hand-typed):

| OVAL document | definition id | rules |
|---|---|---|
| RemoteAccessServicesTest.xml | `oval:org.RemoteAccessServices:def:1` | 46 |
| NoUsersCheck.xml | `oval:org.NoUsers:def:1` | 18 |
| UserPasswordConfiguredTest.xml | `oval:org.example:def:3` | 13 |
| DetectOpenSslTest.xml | `oval:org.OpenSsl:def:1` | 5 |
| VarLogPermissionsTest.xml | `oval:org.varlog:def:2` | 3 |
| LibraryPermissionsTest.xml | `oval:org.LibraryPermissions:def:2` | 2 |
| PackageSignatureTest.xml | `oval:org.PackageSignature:def:1` | 1 |

Each referenced OVAL document defines a single definition, so `name` only
makes explicit what oscap already resolves — **no rule verdict changes**.
The SCE rule (`AslrCheck.sh`) has no OVAL definition and is left as-is;
the CertificateAudit rules already carry a `name`.

## Testing

- `oscap ds sds-validate`: pass.
- Offline matrix: unchanged / green.
- Report generated against `chainguard-base-fips`: totals unchanged
  (92 pass / 107 n/a), and all nine OVAL definitions are now referenced
  in the report (previously only `CABundleHash`), so every evaluated rule
  renders its OVAL test-result detail.

### DIFF
--- a/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
+++ b/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
@@ -3891,7 +3891,7 @@
                     </ns0:description>
           <ns0:ident system="http://cyber.mil/cci">CCI-002450</ns0:ident>
           <ns0:check system="http://oval.mitre.org/XMLSchema/oval-definitions-5">
-            <ns0:check-content-ref href="DetectOpenSslTest.xml"/>
+            <ns0:check-content-ref href="DetectOpenSslTest.xml" name="oval:org.OpenSsl:def:1"/>
           </ns0:check>
         </ns0:Rule>
         <ns0:Rule id="xccdf_mil.disa.stig_rule_SV-203776r959006_rule" selected="true" severity="high">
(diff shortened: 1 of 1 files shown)


xnox:
With this change, openssl checks actually show things they check

image


instead of just being yeah, this is a pass.

egibs (approved):

PR #132, opened 2026-07-23
Replace OpenSSH package ban with FIPS SSH configuration checks

xnox:
## Summary

OpenSSH uses FIPS-validated cryptography, so treating it as a banned remote-access package was incorrect. This PR **drops the OpenSSH ban** and instead **verifies OpenSSH is configured for FIPS**.

- **Un-ban OpenSSH**: remove `openssh`, `openssh-server`, `openssh-client`, `openssh-sftp-server` from the `RemoteAccessServices` banned-package pattern. `dropbear`, `tigervnc`, `xrdp`, `samba`, `telnet`, etc. remain banned.
- **Verify FIPS SSH config** by extending the existing OpenSSL FIPS OVAL definition (`oval:org.OpenSsl:def:1`, which already backs the 5 `Open_Ssl` STIG rules) with checks that:
  - the client drop-in `/etc/ssh/ssh_config.d/10-ssh-fips.conf` and server drop-in `/etc/ssh/sshd_config.d/10-sshd-fips.conf` exist and pin the FIPS-approved `Ciphers` / `KexAlgorithms` / `MACs` / `RequiredRSASize` (plus the server `HostKey`/`CASignature`/`HostBasedAccepted`/`PubkeyAccepted` algorithm lists);
  - the top-level `ssh_config` / `sshd_config` `Include` their respective `*_config.d/*.conf` drop-in directories, so the policy is actually sourced.

### Package-gated
The SSH checks use OVAL implication (nested `criteria operator="OR"` with a negated package-presence test), so:
- **client** checks apply only when `openssh-client` is installed;
- **server** checks apply only when `openssh-server` is installed.

A client-only image such as `cgr.dev/chainguard-private/go-fips` (has `openssh-client`, no `sshd_config`) therefore **passes** — its server policy is not asserted.

All OVAL changes are mirrored in both the standalone `DetectOpenSslTest.xml` and the embedded copy in `ssg-chainguard-gpos-ds.xml`; rule and manual-verification descriptions were updated to match.

## Testing

- **Offline matrix** (`tests/oscap-offline`): RemoteAccessServices fixtures switched off `openssh` onto still-banned `dropbear`, plus a new openssh-now-allowed pass case; new `DetectOpenSsl` cases cover drop-in presence/content, the `Include` wiring, and the client-only / server-only / no-openssh gating. **All verified against the real `oscap` scanner** (`OSCAP_OFFLINE_REQUIRE=1`).
- **e2e**: `remote-access-violation` now installs `dropbear` instead of `openssh-server`.
- Datastream validates with **no new schematron errors** vs. base.
- Scanning the real `go-fips` image now reports all 5 `Open_Ssl` rules **pass** (client policy enforced + correct; server policy gated off).


jimmy-ungerman:
@xnox @stevebeattie I took the suggestions laid out in https://linear.app/chainguard/issue/PSEC-1747/review-stigs-pr-132-replace-openssh-package-ban-with-fips-ssh and now we'll only allow openssh-client for now while we explore the additional checks needed for openssh and openssh-server.

Tested with `go-fips:latest` that contains just the openssh-client and it now passes.

xnox:
@jimmy-ungerman conflicts

jimmy-ungerman:
@xnox fixed

aburrage-chainguard:
**1. The FIPS config checks can pass on a non-FIPS SSH config (highest impact)**

  obj:9–obj:20 just regex for ^\s*Ciphers <approved list>\s*$ somewhere in the drop-in. That's not the same as "the effective policy is FIPS." Four ways to pass while running weak crypto:
  - Host/Match scoping ignored. ^\s* deliberately allows leading whitespace (the client fixture is indented under a Host block). So a drop-in containing Host bastion.internal + indented Ciphers <approved> passes — while every other host falls through to defaults. Same server-side with Match User ....                                                                                
  - Lower-numbered drop-ins win. OpenSSH uses first-obtained-value. 10-ssh-fips.conf leaves 00-–09- free; a 05-legacy.conf with Ciphers 3des-cbc silently overrides and the scan still reports pass.
  - Main-config precedence. tst:21/tst:22 only assert that an Include line exists — not that it precedes any conflicting
  directive. Ciphers aes128-cbc on line 1 of sshd_config with Include on line 2 beats the drop-in, and this passes.
  - No check that the main configs lack conflicting Ciphers/MACs/KexAlgorithms lines.

  Given these are drop-ins the image ships, real-world exploitability is low today, but the check doesn't actually enforce what the rule description claims it enforces.

  **2. The package gate fails open**

  <criterion ... negate="true"/> on tst:23/tst:24 means any reason the package test doesn't evaluate true — apk db unreadable, path differs, package renamed — skips all SSH checks and yields pass. Also ^P:openssh-client(-\d+\.\d+\.\d+)?$ is narrower than the RemoteAccessServices pattern ((-\d[\d.]*(-[a-z][a-z0-9-]*)?)?); harmless today since apk P: lines carry no version, but the two should match. Consider inverting to "config is correct OR package genuinely absent" with an explicit absence test rather than a negated presence test.

xnox:
>   **1. The FIPS config checks can pass on a non-FIPS SSH config (highest impact)**
> 
>   obj:9–obj:20 just regex for ^\s*Ciphers <approved list>\s*$ somewhere in the drop-in. That's not the same as "the effective policy is FIPS." Four ways to pass while running weak crypto:
>   - Host/Match scoping ignored. ^\s* deliberately allows leading whitespace (the client fixture is indented under a Host block). So a drop-in containing Host bastion.internal + indented Ciphers <approved> passes — while every other host falls through to defaults. Same server-side with Match User ....                                                                                
>   - Lower-numbered drop-ins win. OpenSSH uses first-obtained-value. 10-ssh-fips.conf leaves 00-–09- free; a 05-legacy.conf with Ciphers 3des-cbc silently overrides and the scan still reports pass.
>   - Main-config precedence. tst:21/tst:22 only assert that an Include line exists — not that it precedes any conflicting
>   directive. Ciphers aes128-cbc on line 1 of sshd_config with Include on line 2 beats the drop-in, and this passes.
>   - No check that the main configs lack conflicting Ciphers/MACs/KexAlgorithms lines.
> 
>   Given these are drop-ins the image ships, real-world exploitability is low today, but the check doesn't actually enforce what the rule description claims it enforces.
> 
>   **2. The package gate fails open**
> 
>   <criterion ... negate="true"/> on tst:23/tst:24 means any reason the package test doesn't evaluate true — apk db unreadable, path differs, package renamed — skips all SSH checks and yields pass. Also ^P:openssh-client(-\d+\.\d+\.\d+)?$ is narrower than the RemoteAccessServices pattern ((-\d[\d.]*(-[a-z][a-z0-9-]*)?)?); harmless today since apk P: lines carry no version, but the two should match. Consider inverting to "config is correct OR package genuinely absent" with an explicit absence test rather than a negated presence test.
> 

Note that most of the STIG checks can be avaded, as it only statically checks things, and one can always execute ssh client with a cli config options set to anything one desires.

But if openssl is configured, the non-approved algorithms will simply fail at runtime. Thus if one uses non-approved ciphers, the connectivity will fail. The checks that runtime configs are in effect are validated in the openssl-config-fipshardened tests.

And yes removing / tampering with apk db file is a nice way to bypass these checks today; or copy in ssh without apk adding it.

Alternative approach is no possibly build openssh-fips fork that has all the non-approved algorithms compiled out; then we won't even need to perform config checks.

markusboehme:
> Alternative approach is no possibly build openssh-fips fork that has all the non-approved algorithms compiled out; then we won't even need to perform config checks.

I'm not sure that's needed. As you said, a determined person can always evade those checks and e.g. bring a custom SSH client binary. Compliance checks assume the person running them actually has an interest in a compliant system.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalDefinitions/RemoteAccessServicesTest.xml
```diff
@@ -30,7 +30,7 @@
     <textfilecontent54_object xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" id="oval:org.RemoteAccessServices:obj:6" version="1">
       <path>/usr/lib/apk/db</path>
       <filename>installed</filename>
-      <pattern operation="pattern match">^P:(openssh|openssh-server|openssh-client|openssh-sftp-server|dropbear|tigervnc|tigervnc-server|tigervnc-viewer|xrdp|xorgxrdp|vsftpd|proftpd|webmin|cockpit|cockpit-ws|cockpit-bridge|nfs-utils|samba|samba-server|samba-client|samba-common|rsh|telnet)(-\d[\d.]*(-[a-z][a-z0-9-]*)?)?$</pattern>
+      <pattern operation="pattern match">^P:(openssh|openssh-server|dropbear|tigervnc|tigervnc-server|tigervnc-viewer|xrdp|xorgxrdp|vsftpd|proftpd|webmin|cockpit|cockpit-ws|cockpit-bridge|nfs-utils|samba|samba-server|samba-client|samba-common|rsh|telnet)(-\d[\d.]*(-[a-z][a-z0-9-]*)?)?$</pattern>
```
markusboehme: Need to add `openssh-sftp-server` back after restricting the PR to client-only?

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
```diff
@@ -3219,8 +3219,7 @@
         Ethernet switch). <html:pre>
           Script Verification:
           To manually verify, Ensure none of the following remote access packages exist in the
-          folder /lib/apk/db/ Packages: openssh, openssh-server, openssh-client,
-          openssh-sftp-server, dropbear, tigervnc, tigervnc-server, tigervnc-viewer, xrdp, xorgxrdp,
+          folder /lib/apk/db/ Packages: openssh, openssh-server, dropbear, tigervnc, tigervnc-server, tigervnc-viewer, xrdp, xorgxrdp,
```
markusboehme: Need to put back openssh-sftp-server here, too, and in the hunks below?

markusboehme (commented):
The checks for the OpenSSH client are reasonable and, with respect to depth, match what is done in the STIGs for other distributions (well, RHEL and Amazon Linux--Ubuntu and SLES don't have any checks related to the client at all). I'm okay with approving, with the only question being around openssh-sftp-server.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalDefinitions/RemoteAccessServicesTest.xml
jimmy-ungerman: Completely missed that , I'll get it added now.

gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalChecks/XccdfOvalChecks.xml
jimmy-ungerman: Same as above

jimmy-ungerman:
@markusboehme Fixed the `openssh-sftp-server` issues.

markusboehme:
> @markusboehme Fixed the `openssh-sftp-server` issues.

Thanks! Ready to approve from my side once the PR is marked "ready for review".

egibs (approved):

PR #142, opened 2026-08-07
Read trust-store hashes from package stamp files instead of pinning them

xnox:
## Summary

CertificateAudit pinned the SHA-256 of `/etc/ssl/certs/ca-certificates.crt` in the datastream. Every upstream `ca-certificates` roll therefore turned a correctly-updated image into a STIG failure until the daily `update-ca-cert` workflow re-pinned the value and its PR merged — 25+ of the commits in `git log` are exactly that churn.

The `ca-certificates` package already ships the digest next to the bundle, in `sha256sum` format:

```
/etc/ssl/certs/.ca-certificates.crt.sha256   61efbd6d…  ca-certificates.crt
```

So read the expected value from there instead of pinning it. Java images ship the same arrangement for their truststore (`/etc/ssl/certs/java/cacerts` + `.cacerts.sha256`), which the rule previously ignored entirely — that is now covered too, conditionally.

## Changes

**1. CA bundle hash read from the stamp file** (`db34d4a`)

- `obj:4` — `textfilecontent54` extracting the digest via `^([0-9a-fA-F]{64})[ \t]+\*?ca-certificates\.crt$`
- `var:1` — `local_variable` over that subexpression
- `ste:1` — compares `filehash58` against `var:1` rather than a literal
- `tst:4` — new criterion requiring the stamp to exist and parse, so a missing or malformed stamp fails rather than passing vacuously

**2. Java truststore verified where present** (`186a06b`)

Conditional, since non-Java images must not fail for lacking a truststore:

```
OR ─┬─ tst:5  no /etc/ssl/certs/java/cacerts        (non-Java image)
    └─ AND ─┬─ tst:6  stamp exists and parses
            └─ tst:7  cacerts SHA-256 == stamp digest
```

Pairing `tst:6` with `tst:7` is what keeps the OR from being a loophole: a truststore shipped *without* a stamp fails rather than falling through to the absent branch.

**3. Workflow guard on both stamps** (`8cf2a96`)

`update-ca-cert.yaml` no longer has a hash to re-pin, so its extract+`sed` steps are replaced by a `sha256sum -c` guard that fails the run if an image drops a stamp or ships one disagreeing with the file it names — the premise the pin used to guard. It inspects `cgr.dev/chainguard/jre:latest` because that carries **both** stamps, so one export covers the whole rule; `wolfi-base` has no Java and would leave the truststore criterion unguarded. `IMAGE_REF` stays `wolfi-base` for fixture re-pinning, which is a separate concern; both images are now resolved and cosign-verified through one helper.

## Tradeoff — please review this specifically

This is **drift detection, not tamper-evidence**. Whoever can rewrite the bundle can rewrite the stamp beside it. What it still catches is a file modified outside the package (the `cabundle-tampered` fixture), since stamps are package-owned and read-only. If the STIG intent ("only approved trust anchors") demands tamper-evidence, the stamp needs corroboration from something the attacker does not control — apk package integrity or the signed index — not a second file in the same directory. Flagging it rather than deciding it.

## Verification

Real images, via `oscap-docker` against the in-tree datastream:

| image | result |
|---|---|
| `chainguard-base` (Java truststore, consistent) | pass |
| `jdk:latest` | pass |
| `chainguard-base` (no Java) | pass — absent branch |
| CA bundle tampered, stamp untouched | fail |
| truststore tampered, stamp untouched | fail |
| CA bundle updated **with** stamp regenerated | pass *(the released datastream fails this — the false positive removed)* |
| truststore updated **with** stamp regenerated | pass |

That second-to-last row is the point of the change: an A/B against the released datastream baked into the `openscap` image shows it failing a correctly-updated image where this one passes.

New offline fixtures: `fail_wrong_stamp_digest`, `fail_malformed_stamp`, `fail_missing_stamp`, `pass_java_truststore`, `fail_java_wrong_stamp`, `fail_java_missing_stamp`. The offline base has no Java, so those three synthesize the pair with `AddFile` — the truststore's real JKS format is irrelevant to a SHA-256 comparison. This needed two new overlay ops (`ReplaceFile`, `RemoveFile`); `Apply`'s writer already handled removal.

Also run: full `go test -race ./...` green; `make test-e2e-baseline-clean` and `make test-e2e-cabundle-tampered` PASS; `oscap xccdf validate` reports the same pre-existing SRC-9/SRC-10 group errors as `main`, no new ones. The workflow guard body was executed against a real `jre` export — clean passes, each drifted or missing stamp fails with an error naming which trust store.

### DIFF
--- a/.github/workflows/update-ca-cert.yaml
+++ b/.github/workflows/update-ca-cert.yaml
@@ -26,7 +26,13 @@ jobs:
       pull-requests: write
       issues: write
     env:
+      # The image the E2E fixtures are pinned to, re-resolved below.
       IMAGE_REF: cgr.dev/chainguard/wolfi-base:latest
+      # The image the stamp-file guard inspects. jre carries BOTH stamps
+      # CertificateAudit reads — /etc/ssl/certs/.ca-certificates.crt.sha256 and
+      # /etc/ssl/certs/java/.cacerts.sha256 — so one image covers the whole rule.
+      # wolfi-base has no Java and would leave the truststore half unguarded.
+      STAMP_IMAGE_REF: cgr.dev/chainguard/jre:latest
       TESTS_PATH: tests
       FIXTURES_GLOB: tests/e2e/fixtures/*/Dockerfile
       TEST_PIN_FILE: tests/oscap-offline/internal/scan/fixtures_test.go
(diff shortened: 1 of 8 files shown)


aburrage-chainguard (approved):
LGTM.

PR #163, opened 2026-08-24
Check the standalone OVAL checks against their copies in the datastream

stevebeattie:
## Why

The same OVAL definitions live in two places, and nothing compared them.
`validate_checks` and `validate_xml` each check one copy against the schema in
isolation, so a change applied to one and not the other passes CI. That is how
`openssh-sftp-server` came to be banned by the standalone `RemoteAccessServices`
check while the shipped datastream ignored it — the bug fixed in #162.

This adds the check that would have caught it. Reverting #162's datastream
change makes it fail, naming the entity and printing both patterns:

```
::error::RemoteAccessServicesTest.xml: functional content differs (1 differing)
    oval:org.RemoteAccessServices:obj:6 differs:
        standalone : .../pattern|...|^P:(openssh|openssh-server|openssh-sftp-server|...
        datastream : .../pattern|...|^P:(openssh|openssh-server|dropbear|...
```

## What it compares, and why the split

**Functional content** — criteria, tests, objects, states, variables — is an
**error**: the two copies would evaluate the same image differently.

**Metadata** is a **warning**. The copies disagree today on a
`<reference source>` reading `Custom` in every standalone file and `CIS` in five
embedded blocks, which changes no verdict. Failing on it would block adoption
behind a content change. `--strict-metadata` promotes it once that is settled.

## Comparison is per entity, not a flat node list

Worth flagging for review, because the obvious approach is wrong and was what I
wrote first. Flattening both copies into one ordered list of nodes fails three
ways:

- a difference that is a permutation or a count change reports nothing
  locatable — just `0 only in the file, 0 only in the datastream`;
- listing two tests in a different order reads as functional drift, although it
  is schema-valid, changes no verdict, and is something the generator is free to
  do — a false positive on the exact axis this check gates;
- a flat list loses which parent a child belongs to, so a relocated element is
  caught only as a side effect of ordering rather than by comparing structure.

Keying on OVAL id and canonicalising each entity recursively (so a child's path
from its entity root is part of its identity) fixes all three. Duplicate entity
ids are reported rather than one silently displacing the other.

The tests pin those properties directly, since none of the three defects is
detectable from outcome-only assertions: sibling reordering, a namespace-prefix
difference, an element duplicated inside an entity, an entity present in only
one copy, an element relocated between entities, and that a finding names its
entity and shows both values — including for a count-only difference, the case
that previously printed nothing. Each was confirmed to **fail** against the
flat-list implementation rather than merely passing against this one.

## The generator is why this matters beyond tidiness

`chainguard-dev/oscap-playground` assembles the datastream when a new SRG version
lands; its `create_extended_component()` reads each standalone file and embeds it
verbatim. So the standalone files are the **source** and the datastream is a
**build artifact**. Between SRG bumps the datastream is hand-edited and the
generator never runs.

Keeping this green is therefore what makes the next regeneration safe: if the
datastream already matches what the in-repo sources produce, rebuilding cannot
silently revert anything. Worth knowing separately: the generator's own input
copies are currently well behind this repo — its `DetectOpenSslTest.xml` is 72
lines against our 273, and its `CertificateAuditTest.xml` still carries a literal
`<hash>` — so regenerating *today* would revert a great deal. Out of scope here.

## Also reported, not fixed

Definitions present only in the datastream, which therefore cannot be evaluated
with `oscap oval eval` and have no input for regeneration: `oval:org.stub:def:1`
and **CertificateAudit**. The missing `CertificateAuditTest.xml` has to be derived
from the datastream rather than copied from the generator's stale version, so it
is left for a follow-up.

## Usage

```bash
make validate_mirrors            # the check
make validate_mirrors_selftest   # its own unit tests
```

Both run on PRs via the offline workflow — pure Python, repo-local, no registry
or network, so no new allowed endpoints. Not added to the `validate` aggregate,
which already fails on the known-broken `validate_xml`; CI invokes it directly.

## Verification

- Unit tests: 6 methods, 10 table cases. Green here; failing against the
  flat-list implementation.
- `make validate_mirrors` green on the real content — `checked 7` files, 0
  errors, the 5 expected metadata warnings.
- `actionlint` issue count unchanged from `main`; `zizmor` no findings;
  `oscap ds sds-validate` rc=0.
- No datastream or fixture changes in this PR; it is tooling only.

Adds a `.gitignore` (the repo had none) for Python bytecode, plus the scan-output
directories the README already described as ignored.

### DIFF
--- a/.github/workflows/offline-tests.yaml
+++ b/.github/workflows/offline-tests.yaml
@@ -70,6 +70,13 @@ jobs:
       # needs no registry access and adds no allowed endpoints. It covers the
       # trust-store guard's failure paths, which the daily update-ca-cert run
       # never reaches because it only ever sees a healthy image.
+      # Repo-local file comparison: no registry, no network, no extra
+      # endpoints. Compares each standalone OVAL check against its copy in the
+      # datastream, which no schema validation covers because each copy is only
+      # ever checked against the schema in isolation.
+      - name: Check OVAL mirrors against the datastream
+        run: make validate_mirrors
+
       - name: Run trust-store sidecar guard self-test
         run: make test-stamps-selftest
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# Scan output written by the e2e harness and the stigviewer-check target.
+tests/e2e/out/
+out/
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,20 @@ validate_xml:
 
 validate: validate_checks validate_xml
 
-.PHONY: validate validate_xml validate_checks
+# Compare each standalone OVAL check against its copy inside the datastream.
+# The standalone files are the source the datastream is generated from
+# (chainguard-dev/oscap-playground embeds each one verbatim), but between SRG
+# bumps the datastream is edited by hand and nothing runs the generator, so the
+# two drift and it is the generated artifact that ships. validate_checks and
+# validate_xml only ever check each copy against its schema in isolation, so
+# neither notices. Differences that would change a scan verdict fail; purely
+# descriptive ones are logged.
+#
+# Runs as part of `make test-offline` too, since that runs the whole module.
+validate_mirrors:
+	cd tests/oscap-offline && go test -count=1 ./internal/mirrors/
+
+.PHONY: validate validate_xml validate_checks validate_mirrors
 
 # End-to-end scan harness. Each fixture under tests/e2e/fixtures/<name>/
 # builds a container image, runs oscap-docker against the in-repo datastream,
--- a/README.md
+++ b/README.md
@@ -124,6 +124,33 @@ against every individual OVAL definition under
 fall back to `cgr.dev/chainguard/openscap:latest-dev` when `oscap` is
 not installed on PATH.
 
+**Mirror consistency** — `make validate_mirrors` compares each standalone
+OVAL check against its copy embedded in the datastream. Schema
+validation cannot catch drift between them, because it only ever checks
+each copy against the schema in isolation; a change applied to one and
+not the other passes. That is not hypothetical — it is how
+`openssh-sftp-server` came to be banned by the standalone
+`RemoteAccessServices` check while the shipped datastream ignored it.
+
+The standalone files are the source the datastream is built from:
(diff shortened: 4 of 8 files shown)


tests/mirrors/check_mirrors.py
```diff
@@ -0,0 +1,285 @@
+#!/usr/bin/env python3
```
egibs: Is there a benefit to introducing Python for this rather than sticking with Go?

stevebeattie: The only minor benefit is that the tooling in oscap-playground that was intended to generate all this stuff is in python as well, but it's probably not worth it, since that hasn't been used to modify or generate things.

stevebeattie:
Ported to Go — thanks, agreed the Python affinity with `oscap-playground` wasn't worth it given that tooling isn't in use.

The comparison now lives in `tests/oscap-offline/internal/mirrors`, with the repository-wide assertion as `TestRepositoryMirrorsMatch`. `make validate_mirrors` runs that package, and `make test-offline` covers it too since that runs the whole module. No Python left in the repo.

Behaviour is unchanged and I checked equivalence against the real content: seven files compared, the same five `<reference source>` differences logged as non-fatal, the same unmirrored definitions listed, and reverting #162's pattern still fails the test naming `oval:org.RemoteAccessServices:obj:6`.

One thing the port had to get right that the Python didn't: **Go surfaces namespace declarations as ordinary attributes.** The two copies write namespaces differently here — the standalone files redeclare the independent-component namespace as the default on each element, the datastream binds it to a prefix — so comparing attributes naively would have reported every independent element as drift. Declarations are excluded and there's a test for exactly that shape, since it's where this port would otherwise have broken silently.

Two of my test expectations were wrong and got corrected while porting: an element relocated between entities yields **two** findings, not one (a new entity appears and an existing one loses a child — a reader wants both), and a datastream containing no OVAL at all is now an error rather than a per-file "no counterpart" finding, since the latter points at the wrong problem.

Also included: a one-line `gofmt` fix to `internal/scan/fixtures_test.go`, which I left unformatted in #162. That surfaced something worth its own change — there is currently **no `gofmt` or `go vet` gate anywhere**, which is why it slipped through. Following up separately.

egibs (approved):

PR #164, opened 2026-08-25
Add the standalone CertificateAuditTest.xml, derived from the datastream

stevebeattie:
## Why

`CertificateAudit` was the only OVAL component the datastream *references* that
had no standalone file. The datastream names `CertificateAuditTest.xml` in both
its component catalog and the rule's `check-content-ref` — a file that did not
exist in the repo. Two consequences:

- it was the one check that could not be evaluated on its own with
  `oscap oval eval`; and
- regeneration had no input for it. `chainguard-dev/oscap-playground` assembles
  the datastream by reading each standalone file and embedding it verbatim, and
  its `component_id_mapping` expects `'CertificateAuditTest.xml': 'certaudit'`
  by that exact name.

## Derived, not copied

The generator's own copy of this file is stale to the point of being harmful:
**2 tests** against the current **13**, a hardcoded `<hash>` literal predating
even the old pinned value, and no sidecar variable, no `/kaniko` criteria and no
Java truststore. Copying it would have reverted the rule to roughly its 2025
shape.

Instead the `<oval_definitions>` element is lifted out of the `certaudit`
extended-component verbatim and dedented, so the standalone file and the
embedded copy are identical **by construction** rather than by transcription.
That matters because nothing but the mirror check in #163 keeps the two in step.

## Verification

| Check | Result |
|---|---|
| `xmllint --noout` | well-formed |
| `oscap oval validate` | rc=0 |
| `oscap oval eval` | runs the definition standalone — the capability that was missing |
| `make validate_checks` | now validates **8** files, exit 0 |
| #163's comparator | `checked 8` (was 7), **0 errors**, no new warnings |

Against #163's comparator, `CertificateAudit` also drops out of the "present
only in the datastream" list, leaving only `oval:org.stub:def:1` — which has no
standalone file either and is not addressed here.

No datastream change: this adds the missing source file to match what the
datastream already declares.

## Relationship to #163

Independent — this is based on `main`, and I verified it against #163's
comparator by extracting the script from that branch rather than by stacking.
Either merge order works: if #163 lands first, CI here exercises the new pair
automatically; if this lands first, #163's `checked` count goes from 7 to 8.

### DIFF
--- a/gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalDefinitions/CertificateAuditTest.xml
+++ b/gpos/xml/scap/ssg/content/ssg-chainguard-xccdf/OvalDefinitions/CertificateAuditTest.xml
@@ -1,0 +1,175 @@
+<?xml version="1.0"?>
+<oval_definitions xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5" xmlns:oval="http://oval.mitre.org/XMLSchema/oval-common-5" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:ind="http://oval.mitre.org/XMLSchema/oval-definitions-5#independent" xmlns:unix="http://oval.mitre.org/XMLSchema/oval-definitions-5#unix" xsi:schemaLocation="http://oval.mitre.org/XMLSchema/oval-definitions-5 http://oval.mitre.org/language/download/schema/version5.10.1/ovaldefinition/complete/oval-definitions-schema.xsd">
+  <generator>
+    <oval:product_name>Custom</oval:product_name>
+    <oval:product_version>1.0</oval:product_version>
+    <oval:schema_version>5.11.1</oval:schema_version>
+    <oval:timestamp>2025-05-27T12:00:00</oval:timestamp>
+  </generator>
+  <definitions>
+    <definition id="oval:org.CABundleHash:def:1" version="1" class="compliance">
+      <metadata>
+        <title>Validate SHA-256 hash of CA bundle</title>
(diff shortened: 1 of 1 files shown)


stevebeattie:
Verified this resolves correctly against the combined datastream and against the consumer that was failing on it. Recording the checks so a reviewer doesn't have to redo them.

## The file is genuinely derived, with no identifier drift

All **33** OVAL ids in the standalone file (`def:1`, `obj:1-11`, `ste:1-5`, `tst:1-13`, `var:1-3`, all under `oval:org.CABundleHash:`) are present in `ssg-chainguard-gpos-ds.xml`. No id exists here that the datastream doesn't already carry.

## The reference resolves

The datastream's check-content-ref names a specific definition:

```
check-content-ref href="CertificateAuditTest.xml" name="oval:org.CABundleHash:def:1"
```

and this file provides `oval:org.CABundleHash:def:1`. Root element and namespace are right too:

```xml
<oval_definitions xmlns="http://oval.mitre.org/XMLSchema/oval-definitions-5" …>
```

with a single `<definitions>` → `<definition>`, which is the shape the chaindag STIG dashboard's `OvalDefinitions` struct unmarshals (it takes `Definitions[0]`).

## What this fixes downstream

The dashboard fetches OVAL documentation per-rule by reconstructing `…/OvalDefinitions/<href>` and has been logging this on every boot:

```
WARN failed to fetch OVAL definition
  url=…/OvalDefinitions/CertificateAuditTest.xml error="status 404"
```

That failure is not cached, so any finding referencing this rule re-fetched the 404 on every request. This makes it resolve.

Worth noting for anyone tracing the same class: the other 404 in that log, `./AslrCheck.sh`, needs no file. It is one of the two `http://open-scap.org/page/SCE` script checks, and the dashboard filters to the OVAL check system, so it never attempts that fetch. The datastream is self-contained for oscap either way — OVAL is embedded via `ds:component`/`ds:extended-component`, so scanning was never affected by either missing standalone file. This only ever affected consumers that re-fetch by href.

## One cosmetic note

The filename and its contents disagree: `CertificateAuditTest.xml` contains only `oval:org.CABundleHash:*` ids. Harmless — the datastream ref is what binds them, and the dashboard ignores the `name` attribute — but someone grepping the OVAL content for "CertificateAudit" will find nothing. Not worth churning the datastream ref to rename.

egibs (approved):

PR #172, opened 2026-09-01
Remove the unused stub OVAL definition from the datastream

stevebeattie:
## Summary

The datastream carries a "Stub OVAL Definition" component, described as being
*"for manual checks that do not have automated OVAL content."* Remove it — it is
unreachable, it would give the wrong answer if it were reached, and the mechanism
it duplicates is both correct and already in use.

## Three reasons

**Unreachable.** Of 198 rules, 91 carry a `check-content-ref` and **none** names
the stub. Nothing has ever evaluated it.

**Wrong answer if it were reached.** Its criterion is a `textfilecontent54` test
over `/dev/null` with pattern `^$` and `check_existence="all_exist"`. Evaluated
directly:

```
$ oscap oval eval stub.xml
Definition oval:org.stub:def:1: false
```

So a rule wired to it reports **fail**. For a check whose entire point is that it
needs manual review, asserting non-compliance is worse than saying nothing.

**Not fixable into correctness.** The verdict a manual check wants is
`notchecked`, and that comes from a rule having *no check at all* — not from a
definition's result. No OVAL definition can produce it. The 107 rules here that
are manual or not-applicable already do exactly that: no `check-content-ref`, a
`<rationale>` instead, reported as `notchecked`.

## Bonus: one fewer validation warning

`oscap xccdf validate` goes **235 → 234**, and the single line that disappears is
`SRC-207-1|oval-def:definition oval:org.stub:def:1` — the stub was the only thing
that warning was about.

## Verification

- Every `xlink:href` in the datastream still resolves — no dangling references.
- `xmllint` clean, `oscap ds sds-validate` rc=0, `make validate_checks` passes.
- The mirror check's "present only in the datastream" list is now **empty**, so
  that report carries signal rather than a permanent caveat.
- Pre-existing and untouched: `./AslrCheck.sh` is referenced by a
  `check-content-ref` without a matching `cat:uri`, on `main` as well as here.

## Upstream

The component-ref is emitted by `oscap-playground`'s `has_stub_oval` metadata
flag. Filed there separately so a future regeneration does not reintroduce it.

### DIFF
--- a/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
+++ b/gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml
@@ -17,7 +17,6 @@
       </ds:component-ref>
     </ds:checklists>
     <ds:checks>
-      <ds:component-ref id="scap_org.open-scap_cref_stub-oval.xml" xlink:href="#scap_org.open-scap_comp_stub-oval.xml"/>
       <ds:component-ref id="scap_org.open-scap_cref_PackageSignatureTest" xlink:href="#scap_org_ecomp_.packagesignaturecheck"/>
       <ds:component-ref id="scap_org.open-scap_cref_LibraryPermissionsTest" xlink:href="#scap_org_ecomp_.librarypermissions"/>
       <ds:component-ref id="scap_org.open-scap_cref_UserPasswordConfiguredTest" xlink:href="#scap_org_ecomp_.nopasswords"/>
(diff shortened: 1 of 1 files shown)


0xDom-S (commented):
🤖: Verdict: APPROVE. This is a 35-line, zero-addition deletion of the unused/unreachable "Stub OVAL Definition" (`oval:org.stub:def:1`) from `gpos/xml/scap/ssg/content/ssg-chainguard-gpos-ds.xml`. The synthesis review independently re-verified every claim in the PR body with its own XML parser: none of the 91 `check-content-ref`s ever named the stub, the catalog never mapped a `cat:uri` to it (so it was unbindable even by name), its criterion could only ever evaluate to `fail` (never the `notchecked` a manual check needs), all `xlink:href`/`cat:uri` references still resolve post-deletion, and `go test -run TestRepositoryMirrorsMatch` passes green with the datastream-only mirror list now empty. Critical Issues: none. Zero in-scope findings at any severity.

None of the findings below anchor to a diff-map line (the diff map only covers `ssg-chainguard-gpos-ds.xml` lines 17-22 and 6252-6257, i.e., the deleted component-ref/component itself); they reference other, unchanged files and are included here as follow-up suggestions and signals rather than inline comments:

**Suggestions (non-blocking, natural follow-ups this PR enables):**

1. tests/oscap-offline/internal/mirrors/mirrors_test.go:398 — `mirrors.Unmirrored` blocks are only `t.Logf`'d while `report.Functional`/`Descriptive` are `t.Errorf`'d (mirrors_test.go:379-390). The stub was this list's only permanent entry; the review confirmed the list is now empty. Flipping this to `t.Errorf` would cheaply guard against `oscap-playground`'s `has_stub_oval` metadata flag silently reintroducing the component on a future regeneration (a risk the PR body itself names).
2. .github/workflows/create-release.yaml:52 — Full-datastream reference-integrity validation is left commented out here, and `mirrors.Check` (tests/oscap-offline/internal/mirrors/mirrors.go:328-331) only fails on the standalone-file-without-datastream-block direction. A check that resolves every `check-content-ref/@href` through `cat:catalog` and fails on any unresolved href would make this class of change mechanically verifiable in CI, and would immediately surface the pre-existing, unrelated `./AslrCheck.sh` mismatch (ssg-chainguard-gpos-ds.xml:6079, :6098 — no matching `cat:uri`; present on `main`, disclosed in the PR body, out of scope for this change).

**Signals from Dismissed Claims:**

1. Multiple agents cited ephemeral, container-local diff-map filenames (e.g. `claude-guard-diffmap-*.json` under randomized mount roots) as corroborating evidence for cross-agent claims; since these paths/suffixes are randomized per container, this produced at least one false accusation of a fabricated artifact path between agents. Reviews should cite repo-relative paths only, never container-local artifact paths, when cross-referencing other agents' claims.
2. tests/oscap-offline/internal/mirrors/mirrors.go:328-331 and mirrors_test.go:379-381 — the mirror gate does catch deletion of any of the 8 embedded OVAL blocks that mirror a standalone file (it would orphan the standalone file, tripping `report.Functional` → `t.Errorf`); it only misses rules with no standalone counterpart, as was true of the stub. Separately, `tests/e2e/fixtures/*/expected.txt` only asserts outcomes for 8 lines across 4 unique rule ids out of 198 rules (`baseline-clean/expected.txt:10`: "Rules not listed here are not asserted on.") — verdict-assertion coverage is thin; a similar future deletion affecting one of the ~180 unasserted, unmirrored rules would not be caught. This substantiates Suggestion 2 above.
3. tests/stamps/run.sh:83-94 and :178 — the changed datastream is in fact a source that reaches a shell sink: `run.sh` parses `<pattern>` text from `textfilecontent54_object` elements in the datastream and feeds it into `grep -Pq -- "${pattern}"` for a trust-store rule. The deleted stub node was exactly this element type. This is not a finding here — selection is id-keyed at run.sh:87-88 and all three requested object ids (`oval:org.CABundleHash:obj:4/6/10`) survive with one occurrence each — but is noted since an earlier taint-lens pass had incorrectly declared this artifact flow-free. A follow-up self-test asserting `oval_pattern` errors on an unknown id (tests/stamps/run.sh:87-88) would lock this invariant against future regressions.

0xDom-S (approved):
Approved

